PR #13, opened 2015-03-07
Ruby auth add user creds

### LINKED ISSUE #4: ADC Support for User Refresh Tokens
The initial implementation of Application Default Credentials is missing one of the 3 credential types: User Refresh Tokens. This is basically the end result of a 3LO flow, and is an important part of the tool integration, as the Cloud SDK writes out this form of credential in the well-known-file location.

The file format to support is this:

{  
  "type": "authorized_user",     
  "client_id": "kflc91.apps.googleusercontent.com",     
  "client_secret": "1/olgReg3YaBQqxm6T",
  "refresh_token": "2/fFAGRNJru1FTz70BzhT3Zg"     
}
The "type" value is to differentiate it from a service account. The other 3 values are needed to get refresh tokens back. This is a common OAuth2 scenario, so there is likely to be an existing object that already supports getting access tokens from these values. This file could be in either the well-known file location or the environment variable location.

For reference, the Java implementation of these features is here:
https://github.com/google/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java
https://github.com/google/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/UserCredentials.java

tbetbetbe:
Closes #4

### DIFF
--- a/lib/googleauth.rb
+++ b/lib/googleauth.rb
@@ -27,8 +27,13 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-require 'googleauth/service_account'
+require 'multi_json'
+require 'stringio'
+
+require 'googleauth/credentials_loader'
 require 'googleauth/compute_engine'
+require 'googleauth/service_account'
+require 'googleauth/user_refresh'
 
 module Google
   # Module Auth provides classes that provide Google-specific authorization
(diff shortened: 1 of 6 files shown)


tbetbetbe:
@anthmgoogle @vsubramani. PTAL

lib/googleauth.rb
```diff
@@ -40,6 +45,28 @@ module Auth
...
+        clz.new(scope, StringIO.new(MultiJson.dump(json_key)))
+      end
+
+      # Reads the input json and determines which creds class to use.
+      def self.determine_creds_class(json_key_io)
+        json_key = MultiJson.load(json_key_io.read)
+        fail "the json is missing the #{key} field" unless json_key.key?('type')
+        svc_account = json_key['type'] == 'service_account'
+        return json_key, ServiceAccountCredentials if svc_account
+        [json_key, UserRefreshCredentials]
```
anthmgoogle: There may eventually be a 3rd type. Recommend checking the type for user auth and failing with an errir if the type is missing or has any other value.

anthmgoogle:
Looks great. Just one small comment on the JSON parsing.

lib/googleauth.rb
tbetbetbe: Done

PR #16, opened 2015-03-11
Adds an implementation of JWT signing to ServiceAccountCredentials

tbetbetbe:


tbetbetbe:
@anthmgoogle @vsubramani  PTAL

lib/googleauth/service_account.rb
```diff
@@ -43,6 +44,8 @@ module Auth
     #
     # cf [Application Default Credentials](http://goo.gl/mkAHpZ)
     class ServiceAccountCredentials < Signet::OAuth2::Client
```
anthmgoogle: Concerns where raised about having a single type magically switch between the two very different authentication modes, particularly as the IDToken is not strictly OAuth2. While the issue is not closed a safe bet would be to make a separate class for IDToken credentials. For Java I'm planning to call this ServiceAccountIdTokenCredentials.

anthmgoogle:
Implementation looks good. Suggest doing this as a separate class.

tbetbetbe:
@anthmgoogle PTAL.
- it has behaviour split into it's own class, named in the convention that has been discussed
- N.B it is merged with  #17, which should be submitted before this

tbetbetbe:
This change now also contains the promotion to from ServiceAccountCredentials to ServiceAccountJwtHeaderCredentials when the scope is set to nil

PR #17, opened 2015-03-17
Ruby auth make scopes optional

tbetbetbe:


tbetbetbe:
@anthmgoogle PTAL

lib/googleauth.rb
```diff
@@ -76,11 +76,11 @@ def self.determine_creds_class(json_key_io)
     # at http://goo.gl/IUuyuX.
     #
     # If supplied, scope is used to create the credentials instance, when it
-    # can applied.  E.g, on compute engine, the scope is ignored.
+    # can applied.  E.g, on google compute engine, the scope is ignored.
```
anthmgoogle: "can be applied" instead of "can applied".

anthmgoogle: Since there are initially only 3 types, perhaps we should be more specific and mention that this is ignored for user credentials also.

anthmgoogle:
Reviewed. Some minor comments.

anthmgoogle:
LGTM

Issue #27, opened 2015-04-30
Add ability to authenticate a service account via environment variables instead of json file


haabaato:
Due to project requirements, I can't store files with tokens on the server, so all credentials have to read from environment variables.  Adding an initializer to ServiceAccountCredentials that accepts private_key & client_email as arguments instead of reading them from a file would be preferable. [#L66](https://github.com/google/google-auth-library-ruby/blob/master/lib/googleauth/service_account.rb#L66)

I can open a PR if the above solution would work.

tbetbetbe:
Please send the PR - I think what you propose should be OK, we can flesh out the details in a PR.

tbetbetbe:
This was resolved in #28

PR #28, opened 2015-05-17
Enables passing credentials via environment variables instead of json file

haabaato:
Addresses issue #27.  I took it a step further than I originally proposed and made the necessary changes to allow `UserRefreshCredentials` to be initialized from environment variables, in addition to `ServiceAccountCredentials` and `ServiceAccountJwtHeaderCredentials`.


tbetbetbe:
The travis build is failing because of rubocop detected style failings.

I think both the warnings are things that can be ignored, so I'm happy to have the style warnings resolved using by updating the .rubocop_todo.yml.  Could you update that and add it to this PR please?

haabaato:
Updated the .rubocop_todo.yml.  Now the ruby 1.9.3 and jruby builds are failing because of keyword arguments in the initializers.  Is this acceptable, or does this project require 1.9 compatibility?

lib/googleauth.rb
```diff
@@ -52,15 +52,36 @@ class DefaultCredentials
...
+        end
       end
 
       # Reads the input json and determines which creds class to use.
       def self.determine_creds_class(json_key_io)
         json_key = MultiJson.load(json_key_io.read)
-        fail "the json is missing the #{key} field" unless json_key.key?('type')
+        key = 'type'
+        fail "the json is missing the #{key} field" unless json_key.key?(key)
         type = json_key['type']
```
zenhao: `type = json_key[key]` ? + `"the json is missing the '#{key}' field"`

lib/googleauth.rb
```diff
@@ -52,15 +52,36 @@ class DefaultCredentials
...
+        else
+          clz = read_creds
+          clz.new(scope: scope)
+        end
+      end
+
+      def self.read_creds
+        env_var = CredentialsLoader::ACCOUNT_TYPE_VAR
+        fail "#{ACCOUNT_TYPE_VAR} is undefined in env" if ENV[env_var].nil?
+        type = ENV[env_var]
```
zenhao:
```
type = ENV[env_var]
fail "#{ACCOUNT_TYPE_VAR} is undefined in env" unless type
```

lib/googleauth/credentials_loader.rb
```diff
@@ -63,11 +71,14 @@ def make_creds(*args)
...
-        File.open(path) do |f|
-          return make_creds(f, scope)
+        if ENV.key?(ENV_VAR)
+          path = ENV[ENV_VAR]
+          fail "file #{path} does not exist" unless File.exist?(path)
+          File.open(path) do |f|
+            return make_creds(json_key_io: f, scope: scope)
+          end
+        elsif service_account_env_vars? || authorized_user_env_vars?
+          return make_creds(scope: scope)
```
zenhao: To be compatible with old logic
```
elsif service_account_env_vars? || authorized_user_env_vars?
   return make_creds(scope: scope)
else
   return
```

lib/googleauth.rb
```diff
@@ -52,15 +52,36 @@ class DefaultCredentials
...
+        case type
+        when 'service_account'
+          ServiceAccountCredentials
+        when 'authorized_user'
+          UserRefreshCredentials
+        else
+          fail "credentials type '#{type}' is not supported"
```
zenhao: Where can I confirm only 'service_account' and 'authorized_user' should be supported?

lib/googleauth/service_account.rb
```diff
@@ -62,8 +62,15 @@ def self.read_json_key(json_key_io)
       #
       # @param json_key_io [IO] an IO from which the JSON key can be read
       # @param scope [string|array|nil] the scope(s) to access
-      def initialize(json_key_io, scope = nil)
-        private_key, client_email = self.class.read_json_key(json_key_io)
+      # def initialize(json_key_io, scope = nil)
```
zenhao: `# def initialize(json_key_io, scope = nil)` can be removed

lib/googleauth.rb
tbetbetbe: At this point on time, only those two types are supported.   The default credentials flow is defined at https://developers.google.com/identity/protocols/application-default-credentials

lib/googleauth/service_account.rb
```diff
@@ -135,8 +142,13 @@ def self.read_json_key(json_key_io)
...
       # @param json_key_io [IO] an IO from which the JSON key can be read
-      def initialize(json_key_io)
-        private_key, client_email = self.class.read_json_key(json_key_io)
+      def initialize(json_key_io: nil)
+        if json_key_io
+          private_key, client_email = self.class.read_json_key(json_key_io)
+        else
+          private_key = ENV[CredentialsLoader::PRIVATE_KEY_VAR]
+          client_email = ENV[CredentialsLoader::CLIENT_EMAIL_VAR]
+        end
```
zenhao: This is more readable to me, with less if/else branches. Set default values from ENV first. But whether to change is up to you.
```
  private_key = ENV[CredentialsLoader::PRIVATE_KEY_VAR]
  client_email = ENV[CredentialsLoader::CLIENT_EMAIL_VAR]
  private_key, client_email = self.class.read_json_key(json_key_io) if json_key_io
```

zenhao: What are the differences from this?
```
def initialize(json_key_io: nil, scope: nil)
```

lib/googleauth.rb
tbetbetbe: For now, Ruby 1.9.3 and JRuby support are required.

lib/googleauth.rb
haabaato: `key` wasn't defined before, resulting in:
`#<NameError: undefined local variable or method 'key' for Google::Auth::DefaultCredentials:Class>`

So I defined it here under the assumption that key was meant to be 'type' originally.

lib/googleauth/service_account.rb
haabaato: Can't shorten to one line because rubocop enforces the 79 character limit.

lib/googleauth.rb
haabaato: Initializer arguments are now passed as a hash, which should fix support for Ruby 1.9 & JRuby.

lib/googleauth.rb
zenhao: `type = json_key['type']` -> `type = json_key[key]` and `"the json is missing the #{key} field"` to `"the json is missing the '#{key}' field"`

haabaato:
@tbetbetbe Build is green.  Anything left to address in this PR?

sqrrrl:
@tbetbetbe Curious - why still require 1.9.3 support? It's officially past EOL as far as ruby-lang.org goes (expired back in Feb.) Is it for jruby which is still lagging on 2.0+ language support?

haabaato:
Pinging @tbetbetbe.  If there's more to be done, I have some time this weekend.

tbetbetbe:
sorry, I was out for much last week.  I will look through this again and take action by wednesday - sorry about the delay.

haabaato:
thanks for the merge!

haabaato:
One more question: when will the gem be version bumped?

PR #30, opened 2015-06-15
relax the constraint of dependent version of multi_json

igrep:
see also: https://github.com/intridea/multi_json/blob/master/CHANGELOG.md#1111


igrep:
I signed it!

PR #35, opened 2015-06-29
Fix a failing test introduced in #28.

mr-salty:
The environment variables in the 'fails if env vars are set' test
were not actually being set because they used a nonexistent hash key
(which has the effect of unsetting the environment variable).

### DIFF
--- a/spec/googleauth/get_application_default_spec.rb
+++ b/spec/googleauth/get_application_default_spec.rb
@@ -195,6 +195,8 @@ describe '#get_application_default' do
         client_secret: 'privatekey',
         refresh_token: 'refreshtoken',
         client_id: 'app.apps.googleusercontent.com',
+        private_key: @key.to_pem,
+        client_email: 'app@developer.gserviceaccount.com',
         type: 'not_known_type'
       }
     end


mr-salty:
@haabaato PTAL - this makes the test pass but let me know if there's a better fix.

haabaato:
Apologies for the delay as I'm currently attending a conference. :+1: Seems like a simple fix, but I'm unsure as to why this test wasn't previously failing.  I'll take a closer look later today.

mr-salty:
this was addressed more properly in #36 so closing this.

haabaato:
@mr-salty #36 doesn't address this issue. Your PR seems fine to me.

mr-salty:
argh, sorry, I read that too fast.

tbetbetbe:
Is this still necessary ?

mr-salty:
The test still fails for me without this change, so I think it's still needed.

tbetbetbe:
Can you describe a bit more the environment where this test is failing; at the moment Travis is green ?

tbetbetbe:
Ok, this is safe to add as it does not break the tests, and IUC, it breaks your other PR.

mr-salty:
That is a good question about why it was failing for me locally but not failing in travis (which it clearly is not).  I was using ruby 1.9 but that doesn't seem to be it, I also see the failure with 2.2.  @haabaato - any thoughts?

FWIW, here is the (edited) log:

```
$ git checkout caed076711de9375772123560837975f118238fb
[...]
$ ruby --version
ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-linux]
$ rvm --version
rvm 1.26.11 (latest) by Wayne E. Seguin <wayneeseguin@gmail.com>, Michal Papis <mpapis@gmail.com> [https://rvm.io/]
$ bundle --version
Bundler version 1.10.5
$ gem --version
2.4.6
$ bundle exec rake
[...]
Failures:

  1) #get_application_default when credential type is unknown fails if env vars are set
     Failure/Error: expect(&blk).to raise_error RuntimeError
       expected RuntimeError but nothing was raised
     # ./spec/googleauth/get_application_default_spec.rb:240:in `block (3 levels) in <top (required)>'

Finished in 6.48 seconds (files took 0.34353 seconds to load)
82 examples, 1 failure

Failed examples:

rspec ./spec/googleauth/get_application_default_spec.rb:234 # #get_application_default when credential type is unknown fails if env vars are set
```

haabaato:
The test was always passing for me locally, but as I was trying to mention above, it was only passing because `NOT_FOUND_ERROR` is raised in `get_application_default`: https://github.com/google/google-auth-library-ruby/blob/master/lib/googleauth.rb#L116.
Though I had written the test assuming that `read_creds` would raise the error here: https://github.com/google/google-auth-library-ruby/blob/master/lib/googleauth.rb#L77

mr-salty:
Ah, thanks - now I understand exactly what is happening.  In my case it was picking up ~/.config/gcloud/application_default_credentials.json off the local disk and not throwing the exception.  If I remove that file, then the test passes even without this PR.

Using fakefs everywhere (which Tim already opened as #40) would prevent that from happening.  FWIW, I did try doing that (it's literally a 1-liner, 'require fakefs'), but there were some failures, so I went with a more narrow approach for my new tests.  But, it's definitely a good idea. Maybe we can also do ENV.clear to prevent the user's environment from leaking in?

PR #52, opened 2015-11-13
3LO - Installed app support

sqrrrl:
Authorizer tailored to scripts/command line apps, ported from google-apis-ruby-client.

Strictly speaking, doesn't *need* to be a separate authorizer. OOB is fairly trivial to do with the plain UserAuthorizer and the embedded server part could be its own thing. But this does make it relatively trivial to implement CLIs that can switch modes easily...

```ruby
credentials = authorizer.get_credentials(options[:user] || 'default') do |auth|
  if options[:headless]
    puts "Open #{auth.oob_auth_url} to authorize and enter the provided code below:"
    gets
  else
    auth.authorize_with_local_server
  end
end
```


sqrrrl:
@tbetbetbe ping

PR #54, opened 2015-12-09
Prep for 0.5 release

sqrrrl:
Updates to release 0.5 for the user authorizer functionality.

Includes updates to tests to use webmock instead of Faraday stubs. Using faraday's stubs turns out to be fragile as data isn't fully normalized (content of env changes depending on the style of the call, even if they're semantically equivalent.) Using webmock ends up being simpler and safer anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ rvm:
   - jruby
 matrix:
   allow_failures:
-    - rbx-2 # See rubinius/rubinius#3485 - rubocop segfaults
+    - rvm: rbx-2 # See rubinius/rubinius#3485 - rubocop segfaults
 script: "bundle exec rake"
 addons:
   apt:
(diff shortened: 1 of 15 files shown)


tbetbetbe:
LGTM - thanks!

sqrrrl:
RBX error appears unrelated (seems a lot of people have issues with rubocop being flakey with rubinius - should just be disabled IMHO)

And not going to sweat a -.06% change in code coverage :)

tbetbetbe:
LGTM.

rbx-2 should not trigger a failure; I looks like the [.travis.yml](https://github.com/google/google-auth-library-ruby/blob/master/.travis.yml) should stop that, but does not.  I think the allow_failures should be:

```
matrix:
  allow_failures:
    - rvm: rbx-2 # See rubinius/rubinius#3485 - rubocop segfaults
```

rather than 

```
matrix:
  allow_failures:
    - rbx-2 # See rubinius/rubinius#3485 - rubocop segfaults
```

Can you add that to this PR and see if it makes a difference ?

sqrrrl:
Thanks, that did the trick.

PR #56, opened 2015-12-31
Fix for ADC not working on some windows machines

vsubramani:
Fix for https://github.com/google/google-auth-library-ruby/issues/55

Currently only some of the possible host_os values (i.e. Windows and mswin) are used to identify whether the library is running on a windows machine. There are many other possible values that could signify a windows machine. E.g. my machine has 'mingw32'. Using the OS.windows? seems like a more reliable way to identify this.

Please note: this fix is for representational purposes only. I have validated the code fragments individually on my machine, but I haven't validated it by building this full repo.


vsubramani:
Signed the CLA :)

PR #57, opened 2016-01-05
Use lowercase :authorization for metadata key

murgatroid99:
This makes the library HTTP2-compatible without breaking HTTP/1.1 support.

### DIFF
--- a/lib/googleauth/signet.rb
+++ b/lib/googleauth/signet.rb
@@ -32,7 +32,7 @@ require 'signet/oauth_2/client'
 module Signet
   # OAuth2 supports OAuth2 authentication.
   module OAuth2
-    AUTH_METADATA_KEY = :Authorization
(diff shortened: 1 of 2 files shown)


tbetbetbe:
Please confirm that you have manually tested this change.

At a minimum
- do a local install of https://github.com/google/google-api-ruby-client
- update it to use this PR
- successfully complete a call to a Google HTTP/1.1 backend

murgatroid99:
I used the modified library to successfully run the PubSub example in the API client. However, as far as I can tell, it doesn't even use the `Client#apply` method, where that change is relevant.

murgatroid99:
Can we also publish this change to rubygems soon?

tbetbetbe:
if you send another PR with the version bump, it'll be quicker ;)

tbetbetbe:
also update the CHANGELOG as well...

PR #61, opened 2016-01-21
Add Service Accounts section with example for README

sunboshan:


sqrrrl:
The preference is to use `get_application_default` to accomplish that. It does the same thing with the added benefit of leveraging features of GCE to get VM's identity, picking up config from the environment, etc. without having to change any code in your app.

Of course the credentials can be directly instantiated if that's what you really want to do or if there is some constraint that requires it, but I don't think it should be encouraged.

What prompted you to make this PR? If there was a use case that wasn't covered with `get_application_default` then that's something to look at as a separate issue. If it's confusion about the docs/terminology (e.g. unclear that ADC == service account) then that's something that needs to be fixed.

sunboshan:
I think it's unclear that ADC == service account. In the docs for [ADC](https://developers.google.com/identity/protocols/application-default-credentials), I thought [Service Account](https://developers.google.com/identity/protocols/OAuth2ServiceAccount) is a separate thing than ADC. The example in README include *[web](https://github.com/google/google-auth-library-ruby#example-web)* and *[command line](https://github.com/google/google-auth-library-ruby#example-command-line)*, maybe add a *service account* example would make it more clear that service account is ADC.

PR #70, opened 2016-03-23
Use newest token cred uri

unlimit:
Accordingly to [this doc](https://developers.google.com/identity/protocols/OAuth2ServiceAccount#authorizingrequests) token cred uri should be - https://www.googleapis.com/oauth2/v4/token

These changes fix access issues for some Google API's (http://stackoverflow.com/questions/35623453/cant-access-marketplace-license-api-with-service-account)


hxiong388:
@unlimit Thanks for pointing out the issue and submit this PR. I have applied the update from this PR into all the other places that are required in PR #112.

PR #76, opened 2016-07-13
added ruby 2.3 to travis w/ other small changes.

shicholas:
If this project is still actively maintained, I added the latest ruby to the matrix build.

also:

* updated rubocop and applied styling
* made small test refactorings (using the expect block notation)
* made a small refactor to remove an unnecessary ``return credentials``
* bumped the version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: ruby
 rvm:
+  - 2.3
   - 2.2
   - 2.0.0
   - 2.1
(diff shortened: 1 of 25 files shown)


shicholas:
on a tangent Ruby 1.9.3 is now deprecated: https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/ should that still be part of the build matrix?

tbetbetbe:
Hi @shicholas, thanks for this contribution!

The project is actively maintained.  I think I need to check with a few other people before removing 1.9.3 from the build matrix, so it probably won't be part of this change.

PR #77, opened 2016-07-14
Fix travis

shicholas:
copied the file from https://github.com/google/google-api-ruby-client/blob/master/.travis.yml, I hope this works.


hxiong388:
Thanks for this PR, but I have incorporated numerous these build related PRs into #109.

[duplicate of #109 ]

Issue #99, opened 2017-04-01
Faraday Newer Version Support


jwg2s:
Any plans to support newer versions of `faraday`?  Starting to run into gem conflicts that are stuck on v0.9x of `faraday`.

hxiong388:
closed via #100

PR #98, opened 2017-02-25
Fix rubocop warning

serihiro:
# What will this PR change?
- fix all rubocop warning to work Travis CI properly
    - All Travis jobs will be **failed** for rubocop 0.46 warning, for now 😱 
- specify ruby 2.3.3 version in `travis.yml` to fix rvm install error.

## before
```sh
$ bundle exec rake
Running RuboCop...
Inspecting 33 files
CC...........C..C.C.......C......
```

## after
```sh
$ bundle exec rake
Running RuboCop...
Inspecting 33 files
.................................
```

PR #100, opened 2017-04-11
Update faraday dependency

icco:
Faraday 0.9 is two years old. Reading through https://github.com/lostisland/faraday/releases, this should probably be updated.

This is an attempt to deal with #99.


icco:
This looks to only be failing with rubocop changes?

jdelStrother:
Yes please, would love to see this merged

franklinyu:
I think the Rubocop can be fixed in #98. This project has been inactive for almost a year though (so I suspect that it is abandoned).

PR #107, opened 2017-06-30
Avoids redundant class method definition

rthbound:
Adds module to provide `read_json_key` class method to multiple classes.


JustinBeckwith:
@TheRoyalTnetennba please take a look, make sure this all looks right, and merge if it's good :)

PR #157, opened 2018-08-15
Fix auth token revocation (broken since 0.6.3)

tadast:
Looks like https://github.com/google/google-auth-library-ruby/pull/147/files has changed the auth [revocation URL](https://github.com/google/google-auth-library-ruby/blob/43d175f65847e7235e4a7964446443d3b42ea20f/lib/googleauth/user_refresh.rb#L51) from `https://accounts.google.com/o/oauth2/revoke` to `https://oauth2.googleapis.com/revoke` and the new URL seems to respond with a 404. This is because [/revoke should be invoked with a POST](https://github.com/GoogleCloudPlatform/google-cloud-common/issues/260#issuecomment-413232751).

Update The HTTP method for /revoke to be a POST.

### DIFF
--- a/lib/googleauth/user_refresh.rb
+++ b/lib/googleauth/user_refresh.rb
@@ -94,7 +94,7 @@ module Google
         c = options[:connection] || Faraday.default_connection
 
         retry_with_error do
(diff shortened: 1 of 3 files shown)


tadast:
I signed it!

Issue #192, opened 2019-02-05
"No such file or directory - gcloud" warning on Ruby 2.6.1


morgoth:
I updated Rails app to Ruby 2.6.1 and when loading this lib code, I get warning in stdout output: `No such file or directory - gcloud`

This is because of this: https://github.com/googleapis/google-auth-library-ruby/blob/master/lib/googleauth/credentials_loader.rb#L176-L177 when process tries to run `gcloud` command on my system.

Looks like Ruby < 2.6.0 just silently swallowed this output message

I'm using googleauth 0.8.0

TheRoyalTnetennba:
What OS are you using?

morgoth:
It's Ubuntu 18.10
`Linux tp 4.18.0-13-generic #14-Ubuntu SMP Wed Dec 5 09:04:24 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`

TheRoyalTnetennba:
The only way I was able to reproduce this is by passing the --debug argument. Can you post the output of  `echo $RUBYOPT`?

morgoth:
Inside Rails app it's "-rbundler/setup".
I can see a difference in irb and rails console:
```
irb
irb(main):001:0> require 'googleauth'
=> true
irb(main):002:0> Google::Auth::CredentialsLoader.load_gcloud_project_id
=> nil
irb(main):003:0> exit
bin/rails c
Running via Spring preloader in process 17860
Loading development environment (Rails 5.2.2)
[1] pry(main)> Google::Auth::CredentialsLoader.load_gcloud_project_id
rails_console: No such file or directory - gcloud
=> nil
```
Not sure what exactly is happening, but there was no such output when using Ruby 2.6.0

naerymdan:
I also have this exact issue, but it seems to be less about the ruby version rather a regression in googleauth 0.7+:

On `googleauth 0.6.7` I can run google libraries like `google-cloud-storage` and `google-cloud-pubsub` on a Debian or Alpine container on GKE and get implied credentials.

Starting from 0.7 and still in 0.8, I cannot get credentials in my containers, and it logs me `bin/rails: No such file or directory - gcloud` or `bin/rake: No such file or directory - gcloud` by hitting the exact `load_gcloud_project_id` function that tries to call the gcloud executable, which obviously does not exist in a docker container.

morgoth:
[...]

PR #193, opened 2019-02-05
check if gcloud command exists before calling it

TheRoyalTnetennba:
* Fixes https://github.com/googleapis/google-auth-library-ruby/issues/192
* Add double-quotes to linter rules and run `rubocop --auto-correct`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,4 +25,5 @@ Style/PercentLiteralDelimiters: # Contradicting rule
   Enabled: false
 Style/SymbolArray: # Undefined syntax in Ruby 1.9.3
   Enabled: false
-
+Style/StringLiterals:
+  EnforcedStyle: double_quotes
--- a/Gemfile
+++ b/Gemfile
@@ -1,22 +1,22 @@
-source 'https://rubygems.org'
+source "https://rubygems.org"
 
 # Specify your gem's dependencies in googleauth.gemspec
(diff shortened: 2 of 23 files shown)


quartzmo:
Just curious, is there any case to be made for emitting a warning if `gcloud` is not present?

TheRoyalTnetennba:
> Just curious, is there any case to be made for emitting a warning if `gcloud` is not present?

I kind of preferred giving some sort of warning. It was removed in https://github.com/googleapis/google-auth-library-ruby/pull/184. :man_shrugging:

quartzmo:
@TheRoyalTnetennba I remember now that the warning was a sore point.

blowmage:
Changing the linter to double quotes seems orthogonal to adding a check for `gcloud` before calling it. Can the linter change be split into a separate PR?

morgoth:
This fixes the case when `gcloud` is not installed.
However on Travis when I run tests, there is still output displayed:
```
ERROR: (gcloud.config.config-helper) You do not currently have an active account selected.
Please run:
  $ gcloud auth login
to obtain new credentials, or if you have already logged in with a
different account:
  $ gcloud config set account ACCOUNT
to select an already authenticated account to use.
```
I'm running tests on Travis using `xenial` distro

morgoth:
@TheRoyalTnetennba any other idea how to fix it?

PR #203, opened 2019-03-14
add google-style and refactor

TheRoyalTnetennba:


blowmage:
Ubuntu looks like it is running 4 ruby versions, but OSx and Windows are only running 1 ruby version. Is that intentional?

TheRoyalTnetennba:
> Ubuntu looks like it is running 4 ruby versions, but OSx and Windows are only running 1 ruby version. Is that intentional?

I have no control over the OSx environment. WIndows, I'm working on.

y-yagi:
This has changed the required Ruby version. Ref: #208.
If this minimum version of the change is intentional, I think it should define `required_ruby_version` in the gemspec. 
It helps to find the appropriate gems.

yahonda:
Updated #208 to define ` required_ruby_version` https://github.com/googleapis/google-auth-library-ruby/pull/208/commits/7c4209e9f397b1ef844916c175211fb47ef8de90 .

PR #215, opened 2019-05-15
Fix the gcloud command to avoid spamming stderr when unauthorized.

igorpeshansky:
Without this change, creating the default credentials when gcloud credentials are not set results in the following on stderr:
```
$ ruby -e 'require "googleauth"; Google::Auth::Credentials.default'
ERROR: (gcloud.config.config-helper) You do not currently have an active account selected.
Please run:

  $ gcloud auth login

to obtain new credentials, or if you have already logged in with a
different account:

  $ gcloud config set account ACCOUNT

to select an already authenticated account to use.
```
See also googleapis/google-cloud-ruby#2986.


qingling128:
@TheRoyalTnetennba - Any chance we know when the next release of `google-auth-library-ruby` would be?

PR #275, opened 2020-10-02
Fix passing multiple scopes to metadata server.

danielroseman:
In #264, a fix was made to allow passing the scopes in the request to
the metadata server. A comment on that PR pointed to Signet's use of a
space-delimited string for muliple values, so the same was implemented
here.

However this code does not actually use Signet to make the request. In
fact the server is expecting a standard form-encoded parameter, eg
`scopes=scope1&scopes=scope2`. Passing a space results in a 500 response
from the server.

This change corrects the request to use the form-encoded format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Release History
 
+### Unreleased
+
+* Fix issue with passing multiple scopes in metadata server authentication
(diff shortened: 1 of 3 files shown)


danielroseman:
Fixed in #277.

PR #305, opened 2021-03-26
Accept application/text content-type for plain idtoken response

bouk:
On Cloud Run, the idtoken metadata endpoint has Content-Type application/text. The current code makes it impossible to fetch an id token, because the raw JWT gets passed to parse_credentials which then fails because it doesn't recognize the content-type


dazuma (approved):
fix: Accept application/text content-type for plain idtoken response

PR #336, opened 2021-09-22
Add `ServiceAccountCredentials.from_env` example

MarkBennett:
A quick documentation update to add an example of creating a service worker auth from the environment


dazuma (approved):
Thanks!

PR #338, opened 2021-09-24
CredentialsLoader: close stdin and stderr to prevent reauth

bouk:
CredentialsLoader shells out to gcloud which causes a reauth prompt to
appear if the user's credentials have been expired. We should close
stdin and stderr so the shelling out silently fails and we return nil
instead.

If we don't do this we can randomly get prompts as follows:

```
Reauthentication required.
Please enter your password:
```


dazuma (approved):
Thanks!

PR #344, opened 2021-10-07
Add instructions for 3-legged-oauth

jcmorrow:
I spent a *long* time figuring out how to do this properly. The docs on
domain-wide delegation don't really make it clear that `sub` has to
be set in order to take advantage of those settings. It seems to me like
this is the easiest way to run domain-wide delegation with this gem, but
if there's an easier way I'm all ears!


SandeepTuniki:
Hi @jcmorrow, thanks for raising this PR and apologies for delay in response. The changes look good to me. Merging it now.

SandeepTuniki (approved):
docs: Add README instructions for 3-Legged OAuth with a service account

Issue #351, opened 2021-11-11
User Credentials > Example (Command Line) undefined local variable: user_id


simplytech:
I'm trying to follow the README.md instructions for User Credentials > Example (Command Line)

I have a command line app (CLI) that runs on the client and needs the end-user to authorize it to access the end-user email. I understand that the OAuth 2.0 Scope for Google API would be:

https://www.googleapis.com/auth/userinfo.email

but for the sake of this discussion, let's just use the scope for Drive as in the example.

It seems that the code is using ```user_id``` before defining it.

#### Environment details

  - OS: macOS Version 10.14.6 (18G9323)
  - Ruby version: ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin18]
  - Gem name and version:     googleauth (1.1.0)

#### Steps to reproduce

  1. Download the client_secrets.json from GCP > (my project name) > Credentials > OAuth 2.0 Client IDs (Name is "my_name", Type is "Desktop") and put it in the current directory ```./client_secrets.json```
  2. Copy the code in "Example (Command Line)" and paste into file called ```./example-command-line.rb```
  2. Replace ```/path/to/client_secrets.json``` with ```./client_secrets.json```
  3. Replace ```/path/to/tokens.yaml``` with ```./tokens.yaml```
  2. run the command ```ruby ./example-command-line.rb```

#### Code example

```ruby
# Example (Command Line)
require 'googleauth'
require 'googleauth/stores/file_token_store'

OOB_URI = 'urn:ietf:wg:oauth:2.0:oob'

scope = 'https://www.googleapis.com/auth/drive'
client_id = Google::Auth::ClientId.from_file('./client_secrets.json')
token_store = Google::Auth::Stores::FileTokenStore.new(:file => './tokens.yaml')
authorizer = Google::Auth::UserAuthorizer.new(client_id, scope, token_store)

credentials = authorizer.get_credentials(user_id)
if credentials.nil?
  url = authorizer.get_authorization_url(base_url: OOB_URI )
  puts "Open #{url} in your browser and enter the resulting code:"
  code = gets
  credentials = authorizer.get_and_store_credentials_from_code(user_id: user_id, code: code, base_url: OOB_URI)
end

# OK to use credentials
```

I get the error:
```
Traceback (most recent call last):
example-command-line.rb:12:in `<main>': undefined local variable or method `user_id' for main:Object (NameError)
```

quartzmo:
[...]

PR #352, opened 2021-11-12
docs: Fix sample in README.md

quartzmo:
closes: #351

I think it seems reasonable and simple to initialize `user_id` to `ENV['USER']` for this command-line example, but I am open to other ideas as well, such as using [Etc](https://ruby-doc.org/stdlib-3.0.2/libdoc/etc/rdoc/Etc.html).

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ token_store = Google::Auth::Stores::FileTokenStore.new(
   :file => '/path/to/tokens.yaml')
 authorizer = Google::Auth::UserAuthorizer.new(client_id, scope, token_store)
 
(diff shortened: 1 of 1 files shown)


dazuma (approved):

skarmakar:
@quartzmo `user_id` to `ENV['USER']` is not entirely making it obvious what to put in `ENV['USER']`. Also `USER` makes things a bit confusing whether it's a User object or an identifier. As far I have tested this is the client User identifier.

Maybe `user_id = ENV['USER_ID'] # client  user's identifier` would make it more obvious?

PR #373, opened 2022-03-02
chore: use public method lookup_middleware

NivedhaSenthil:
- load_middleware is made private in later versions of Faraday


bajajneha27 (approved):
LGTM

PR #397, opened 2022-09-19
feat: pass key sequence when creating key

NivedhaSenthil:
- with openssl3 keys are immutable
- update to pass in keys in der format instead of mutating after creating key

### DIFF
--- a/lib/googleauth/id_tokens/key_sources.rb
+++ b/lib/googleauth/id_tokens/key_sources.rb
@@ -130,13 +130,14 @@ module Google
             end
             n_bn = OpenSSL::BN.new n_data, 2
             e_bn = OpenSSL::BN.new e_data, 2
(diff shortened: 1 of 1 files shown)


ariccio:
If you have a moment, could you elaborate on how you construct the sequence here? I know so little about OpenSSL that I don't even know which docs to check.

NivedhaSenthil:
Hi @ariccio, you may refer these docs for info on the structure of keys and more.

RSA: https://www.rfc-editor.org/rfc/rfc3447#appendix-A 
EC: https://www.rfc-editor.org/rfc/rfc5480#section-2

dazuma (approved):
I added tests to verify the key contents, and updated the required tests to match the new CI matrix.

The code LGTM. Thanks Nivedha!

PR #408, opened 2022-12-16
feat: Add AWS Workload Identity Federation Support 

ScruffyProdigy:

PR #414, opened 2023-01-27
feat: Add AWS Workload Identity Federation Support

BigTailWolf:
This is a re-catch on the closed pull request:
https://github.com/googleapis/google-auth-library-ruby/pull/408


BigTailWolf:
@dazuma Would you please take a look at this. We need to have it approved and unblock the customer request.

lib/googleauth/external_account/aws_credentials.rb
```diff
@@ -0,0 +1,346 @@
...
+                       end
+
+            raise Faraday::Error unless response.success?
+            response
+          rescue Faraday::Error
+            raise "Failed to retrieve AWS #{name}."
+          end
+        end
+
+        def retrieve_subject_token!
```
bajajneha27, line 85: Can we please document what these methods do? May be like we have in python library too?

bajajneha27:
@BigTailWolf I just noticed that the CLS is not signed. Can we please fix that?

lib/googleauth/external_account/aws_credentials.rb
```diff
@@ -0,0 +1,346 @@
...
+        # Args:
+        #     region_name (str): The AWS region to use.
+        def initialize region_name
+          @region_name = region_name
+        end
+
+        # Generates the signed request for the provided HTTP request for calling
+        # an AWS API. This follows the steps described at:
+        # https://docs.aws.amazon.com/general/latest/gr/sigv4_signing.html
+        # Args:
```
bajajneha27, line 195: Can we please follow the ruby style guide for documenting the parameters and method description? It helps us keep the public document updated. 
You can find a reference over [here](https://github.com/googleapis/google-api-ruby-client/blob/main/generated/google-apis-storage_v1/lib/google/apis/storage_v1/service.rb#L56)

lib/googleauth/external_account/aws_credentials.rb
```diff
@@ -0,0 +1,346 @@
...
+        #     method (str): The HTTP method used to call this API.
+        # Returns:
+        #     Hash[str, str]: The AWS signed request dictionary object.
+        def generate_signed_request aws_credentials, original_request
+          uri = Addressable::URI.parse original_request[:url]
+          raise "Invalid AWS service URL" unless uri.hostname && uri.scheme == "https"
+          service_name = uri.host.split(".").first
+
+          datetime = Time.now.utc.strftime "%Y%m%dT%H%M%SZ"
+          date = datetime[0, 8]
```
bajajneha27, line 209:
```suggestion
          datetime = Time.now.utc.strftime "%Y%m%d"
```

lib/googleauth/helpers/connection.rb
```diff
@@ -0,0 +1,35 @@
...
+  module Auth
+    # Helpers provides utility methods for Google::Auth.
+    module Helpers
+      # Connection provides a Faraday connection for use with Google::Auth.
+      module Connection
+        module_function
+
+        attr_accessor :default_connection
+
+        def connection
```
bajajneha27, line 29: Do we plan on adding more stuff here?

BigTailWolf:
> @BigTailWolf I just noticed that the CLS is not signed. Can we please fix that?

Hi Neha, I think the security check failed due to Ryan's CLA status. Mine is under Google CLA.
Yet I don't know how to remove Ryan (Scuffypuppy) from the check.

lib/googleauth/external_account/aws_credentials.rb
BigTailWolf: I don't think the following signing part is using `date`. According to my reading of the code. The signing is using the `datetime` which is the `Time.now.utc.strftime "%Y%m%dT%H%M%SZ"

lib/googleauth/external_account/base_credentials.rb
```diff
@@ -0,0 +1,140 @@
...
+          "https://www.googleapis.com/auth/iam".freeze
+        ].freeze
+
+        def token_type
+          # This method is needed for BaseClient
+          :access_token
+        end
+
+        def base_setup options
+          self.default_connection = options[:connection]
```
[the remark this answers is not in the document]
BigTailWolf: In python logic is in the constructor of `STSClient` we are providing a `token_url` and a `client_auth`. The `client_auth` is Nullable. The `client_auth` can be built by `client_id` and `client_secret` if we are given those two parameter in the `options`

Here in the Ruby library I don't quite sure that `Google::Auth::OAuth2::STSClient`'s second parameter's default value. If we can use `Helpers::Connection` as the default. Then we are OK.
To me the understanding of here is we are using `options[:connection]` which is whatever we are given in the initialization list.

lib/googleauth/helpers/connection.rb
BigTailWolf: I believe yes. My understanding here is Ruby Faraday connection is a wrapped layer of Http Request/Response which is almost the same of Python library's `google/auth/transport` package. We are going to support more features like mtls, customer tls, etc.

spec/googleauth/apply_auth_examples.rb
```diff
@@ -21,6 +21,7 @@
 
 shared_examples "apply/apply! are OK" do
   let(:auth_key) { :authorization }
+  let(:extra_checks) {}
```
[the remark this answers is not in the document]
BigTailWolf: Not as now. Maybe more upcoming test examples needs it.

BigTailWolf (commented):
Comments addressed

spec/googleauth/apply_auth_examples.rb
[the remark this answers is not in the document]
bajajneha27: In that case, I think we should remove it for now. And let's add it in upcoming changes.

spec/googleauth/external_account/aws_credentials_spec.rb
```diff
@@ -0,0 +1,511 @@
+# Copyright 2022 Google, Inc.
```
bajajneha27, line 1:
```suggestion
# Copyright 2023 Google, Inc.
```

spec/googleauth/external_account_spec.rb
```diff
@@ -0,0 +1,228 @@
+# Copyright 2022 Google, Inc.
```
bajajneha27, line 1:
```suggestion
# Copyright 2023 Google, Inc.
```

spec/googleauth/oauth2/sts_client_spec.rb
```diff
@@ -0,0 +1,86 @@
+# Copyright 2022 Google, Inc.
```
bajajneha27, line 1:
```suggestion
# Copyright 2023 Google, Inc.
```

bajajneha27 (commented):
Jin, I added a few more nit comments. Otherwise it looks good.

Issue #299, opened 2021-02-24
Getting id-token from compute engine default service account not working


feliara8:
#### Environment details

  - OS: Compute engine with machine for docker containers
  - Ruby version: 2.7.1
  - Gem name and version:  googleauth last version
#### Description

I want to make a call to an api in a iap secured app engine from a compute engine with the proper service account.
I am trying to follow this [link](https://cloud.google.com/iap/docs/authentication-howto#iap_make_request-nodejs)
which included the ruby code to do the job, but its not working

NOTE: Node code works perfectly

#### Steps to reproduce

  1. Create a compute engine with a service account
  2. Go to ruby console
  3. Run ruby code to get id_token, but access token is retrieved despite passing target_audience

#### Code example

```ruby
# url = "The Identity-Aware Proxy-protected URL to fetch"
# client_id = "The client ID used by Identity-Aware Proxy"
require "googleauth"
require "faraday"

# The client ID as the target audience for IAP
id_token_creds = Google::Auth::Credentials.default target_audience: client_id

headers = {}
id_token_creds.client.apply! headers

resp = Faraday.get url, nil, headers

if resp.status == 200
  puts "X-Goog-Iap-Jwt-Assertion:"
  puts resp.body
else
  puts "Error requesting IAP"
  puts resp.status
  puts resp.headers
end
```
Thanks!

bouk:
It seems the target_audience option is not being saved. If you do `id_token_creds.target_audience` you'll get `nil`

ernsheong:
Thanks for workaround, I also expected `Google::Auth::Credentials.default target_audience: ...` to work because of sample code in https://github.com/googleapis/google-auth-library-ruby/pull/258 

But it is also not yet documented:
https://www.rubydoc.info/github/google/google-auth-library-ruby/Google/Auth/Credentials#default-class_method

bkoski:
So many thanks @StupidCodeFactory, I spent an hour banging my head against this before I realized the fix was in the latest gem!

PR #425, opened 2023-04-02
fix: GCECredentials - Allow retrieval of ID token

StupidCodeFactory:
**CONTEXT:** 

While deploying a couple of services on `Cloud RUN`:
1. service A running with service account A
2. service B running with service account B
3. I added `roles/run.invoker` for service account B on service A.

I expected to be able to retrieve an ID token and be able to call service A from service B. After some debugging I realised I was getting an `access_token` rather then an `id_token`.
I tracked it down to the GCECredentials instantiation only passing the scope to the subclass of the [Signet::Oauth2::Client](https://github.com/googleapis/signet/blob/main/lib/signet/oauth_2/client.rb#L95), hence not being able to pass the `target_audience` for retrieve an `id_token`

I believe this should fix: https://github.com/googleapis/google-auth-library-ruby/issues/299

### DIFF
--- a/spec/googleauth/get_application_default_spec.rb
+++ b/spec/googleauth/get_application_default_spec.rb
@@ -125,17 +125,36 @@ describe "#get_application_default" do
       end
     end
 
-    it "succeeds without default file or env if on compute engine" do
-      stub = stub_request(:get, "http://169.254.169.254")
-             .to_return(status:  200,
-                        headers: { "Metadata-Flavor" => "Google" })
-      Dir.mktmpdir do |dir|
-        ENV.delete @var_name unless ENV[@var_name].nil? # no env var
-        ENV["HOME"] = dir # no config present in this tmp dir
-        creds = Google::Auth.get_application_default @scope, options
-        expect(creds).to_not be_nil
+    describe "when on compute engine" do
+      before do
+        @compute_metadata_server = stub_request(:get, "http://169.254.169.254")
+                                     .to_return(status: 200, headers: { "Metadata-Flavor" => "Google" })
       end
-      expect(stub).to have_been_requested
+
+      it "succeeds without default file or env if on compute engine" do
+        Dir.mktmpdir do |dir|
+          ENV.delete @var_name unless ENV[@var_name].nil? # no env var
+          ENV["HOME"] = dir # no config present in this tmp dir
+          creds = Google::Auth.get_application_default @scope, options
+          expect(creds).to_not be_nil
+        end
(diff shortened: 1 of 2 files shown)


spec/googleauth/get_application_default_spec.rb
```diff
@@ -125,17 +125,36 @@
...
+          ENV.delete @var_name unless ENV[@var_name].nil? # no env var
+          ENV["HOME"] = dir # no config present in this tmp dir
+          creds = Google::Auth.get_application_default @scope, options
+          expect(creds).to_not be_nil
+        end
+        expect(stub).to have_been_requested
+      end
+
+      it "honors passing options to OAuth 2 client" do
+        stub = stub_request(:get, "http://169.254.169.254")
```
bajajneha27, line 143: Nit: since we're repeating this stub request. Can you put it in a variable?

bajajneha27 (commented):
Thanks for raising this PR. Just one tiny comment, overall LGTM

spec/googleauth/get_application_default_spec.rb
```diff
@@ -125,17 +125,35 @@
...
-             .to_return(status:  200,
-                        headers: { "Metadata-Flavor" => "Google" })
-      Dir.mktmpdir do |dir|
-        ENV.delete @var_name unless ENV[@var_name].nil? # no env var
-        ENV["HOME"] = dir # no config present in this tmp dir
-        creds = Google::Auth.get_application_default @scope, options
-        expect(creds).to_not be_nil
+    describe "when on compute engine" do
+      before do
+        @compute_metadata_server = stub_request(:get, "http://169.254.169.254")
```
StupidCodeFactory, line 130: @bajajneha27 thanks for feedback.
 While one would _normally_ use a `let!(:compute_metadata_server)` as it's more idiomatic to RSpec test suite, this file seemed to be only be using `@instance_variable` so for the sake of consistency I've kept the pattern used in the file.

There seem to be a mixed use of `let` and `before { @instance_varialbe = "foo" }` throughout the test suite. It's also unclear what is the preferred way.
As a long term rubyist I'd be inclined to favour the `let {}`.
I'd be happy to convert this file from using instance  variable to `let` blocks, in this pull request or another one.

bajajneha27: This looks fine. Thanks for implementing this.

bajajneha27:
@StupidCodeFactory , you'd have to update your branch with main.

bajajneha27 (approved):

PR #427, opened 2023-04-04
fix: Ruby: Remove external account config validation

BigTailWolf:


bajajneha27 (commented):
Changes look good to me but I noticed that the [NodeJs PR](https://github.com/googleapis/google-auth-library-nodejs/pull/1517) and python PR have it as a fix.  Any idea why those are fix?

I also see [`Security Considerations`](https://github.com/googleapis/google-auth-library-nodejs/pull/1517/files#diff-df78b4580f896b4bfa3eba25bc40c45ab1673a5a60f2cac25764a128a0a372afR891) sections in the README of NodeJs PR. I think we should have something like that in our READMEs as well.

BigTailWolf:
The `README.md` change won't be reflect here. Because that's when we finish the pool-based identity and the document would go with that.

bajajneha27 (approved):

PR #437, opened 2023-06-24
feat: Adding support for pluggable auth credentials

BigTailWolf:
Federated OIDC support for workload/workforce in case of the user cannot using a file or url to fetch subject token, we can allow user to provide an executable that we can run to fetch subject token.

### DIFF
--- a/lib/googleauth/external_account/pluggable_credentials.rb
+++ b/lib/googleauth/external_account/pluggable_credentials.rb
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require "open3"
 require "time"
 require "googleauth/external_account/base_credentials"
 require "googleauth/external_account/external_account_utils"
@@ -77,7 +78,7 @@ module Google
           env = inject_environment_variables
           output = subprocess_with_timeout env, @credential_source_executable_command,
                                            @credential_source_executable_timeout_millis
-          response = MultiJson.load output, encoding: "utf-8"
+          response = MultiJson.load output, symbolize_keys: true
           parse_subject_token response
         end
 
--- a/lib/googleauth/external_account/pluggable_credentials.rb
+++ b/lib/googleauth/external_account/pluggable_credentials.rb
@@ -1,0 +1,154 @@
+# Copyright 2023 Google, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "time"
+require "googleauth/external_account/base_credentials"
+require "googleauth/external_account/external_account_utils"
+
+module Google
+  # Module Auth provides classes that provide Google-specific authorization used to access Google APIs.
+  module Auth
+    module ExternalAccount
+      # This module handles the retrieval of credentials from Google Cloud by utilizing the any 3PI
+      # provider then exchanging the credentials for a short-lived Google Cloud access token.
+      class PluggableAuthCredentials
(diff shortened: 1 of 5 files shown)


lib/googleauth/external_account/pluggable_credentials.rb
```diff
@@ -0,0 +1,156 @@
...
+          raise "The executable response is missing the success field." if response[:success].nil?
+        end
+
+        def inject_environment_variables
+          env = ENV.to_h
+          env["GOOGLE_EXTERNAL_ACCOUNT_AUDIENCE"] = @audience
+          env["GOOGLE_EXTERNAL_ACCOUNT_TOKEN_TYPE"] = @subject_token_type
+          env["GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE"] = "0" # only non-interactive mode we support.
+          unless @service_account_impersonation_url.nil?
+            env["GOOGLE_EXTERNAL_ACCOUNT_IMPERSONATED_EMAIL"] = service_account_email
```
bajajneha27, line 136: QQ: why do we call the method `service_account_email` when it fetches impersonated email ?

lib/googleauth/external_account/pluggable_credentials.rb
```diff
@@ -0,0 +1,156 @@
...
+            env["GOOGLE_EXTERNAL_ACCOUNT_OUTPUT_FILE"] = @credential_source_executable_output_file
+          end
+          env
+        end
+
+        def subprocess_with_timeout environment_vars, command, timeout_seconds
+          Timeout.timeout timeout_seconds do
+            output, error, status = Open3.capture3 environment_vars, command
+            unless status.success?
+              raise "Executable exited with non-zero return code #{status.exitstatus}. Error: #{output}, #{error}"
```
bajajneha27, line 148: Do we really need to raise the `output` also here? I think the error message and backtrace should be enough.

lib/googleauth/external_account/pluggable_credentials.rb
BigTailWolf: Yes, we do.

lib/googleauth/external_account/pluggable_credentials.rb
BigTailWolf: Because the user's script is not guarantee to print into stderr. There are a lot of cases people print out their error message into stdout.

lib/googleauth/external_account/pluggable_credentials.rb
bajajneha27: I mean we could rename it so that it reflects that it fetches impersonated email.

BigTailWolf: Oh, we are not ONLY use it when fetching impersonated email. But when we going to impersonate, the `service_account_email` is source of truth.

lib/googleauth/external_account/pluggable_credentials.rb
```diff
@@ -0,0 +1,156 @@
...
+          if @credential_source_executable_timeout_millis < EXECUTABLE_TIMEOUT_MILLIS_LOWER_BOUND ||
+             @credential_source_executable_timeout_millis > EXECUTABLE_TIMEOUT_MILLIS_UPPER_BOUND
+            raise "Timeout must be between 5 and 120 seconds."
+          end
+          @credential_source_executable_output_file = @credential_source_executable[:output_file]
+        end
+
+        def retrieve_subject_token!
+          unless ENV[ENABLE_PLUGGABLE_ENV] == "1"
+            raise "Executables need to be explicitly allowed (set GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES to '1') " \
```
bajajneha27, line 71: Are we not planning to check `allow_executables` ?

bajajneha27:
Do we not need to use `GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE` ? Is it because we still don't have the distinction between interactive mode and non-interactive mode in client libraries?

lib/googleauth/external_account/pluggable_credentials.rb
BigTailWolf: We have to check `GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES` since it's a universal setting from all library.

bajajneha27 (approved):

PR #438, opened 2023-06-26
docs: fixed iss argument and description in comments of IDTokens

corrupt952:
Comments in the `Google::Auth::IDTokens` class that should be `iss` are `aud`.
This PR corrects the comment to `iss` and changes it so that the generated documentation also has the expected description.

Thank you for reviewing.


bajajneha27 (approved):

bajajneha27:
Thanks.

PR #479, opened 2024-03-05
fix: Get the project_id from gcloud

JamshedVesuna:
I think the spirit of https://github.com/googleapis/google-auth-library-ruby/commit/990250345d6af31de1066c08c0b3b42692ae263c was to prevent `gcloud` from popping out of the console and hanging. Unfortunately, this will always close the subprocess and return an empty string.

This causes downstream issues, for example when initializing a new BigQuery client [0] that relies on the credential to carry the `project_id`.

Adding `--quiet` will disable interactive prompts and either attempt to use defaults or error.

0. https://github.com/googleapis/google-cloud-ruby/blob/main/google-cloud-bigquery/lib/google/cloud/bigquery.rb


JamshedVesuna:
Hi, is it possible to get a review for this? I have signed the Contributor License Agreement. Thank you! @dazuma

dazuma (approved):
Sorry we missed this. Yes this looks reasonable.

PR #486, opened 2024-08-12
fix(spec): Set ENV for not enabled spec

blowmage:
The not enabled spec will fail when another spec has set the ENV before the spec is run.


dazuma (approved):
Thanks @blowmage !

PR #487, opened 2024-08-12
refactor: Use NoMethodError for base classes

blowmage:
It is not advisable to raise  NotImplementedError for errors that are not platform errors, such as fsync or fork.
Replace with NoMethodError, which inherits from StandardError.


dazuma (approved):
Good call. Thanks!

PR #519, opened 2025-02-06
fix: `attribute accessor as module_function` warning

byroot:
```
googleauth-1.13.1/lib/googleauth/helpers/connection.rb:27: warning: attribute accessor as module_function
```

This code was introduced in 1354b076a159d223b737e5b1f27dde0463eb5292.

`module_function` cause the next methods to be defined as both class methods and private instance methods, but the class method part is never used, hence directly declaring them as private is simpler and avoid the warning.


viacheslav-rostovtsev:
I don't think `private` is the right way to go here, this is an existing functionality after all. The warning can be dealt with by removing the accessor.

byroot:
> this is an existing functionality after all

Is it? It declare some methods that are useless, they'd set ivars on a module but nothing would ever read it. This `module_function` is very clearly a mistake.

But up to you, as long as this library stops emitting warnings, I'm happy.

PR #545, opened 2025-09-30
fix: Deprecate method make_creds in DefaultCredentials

aandreassa:


lib/googleauth/credentials.rb
viacheslav-rostovtsev, line 503: I think this should be a call to `init_client`. 
And is there a reason that init_client is an instance method instead of `self.` method?

lib/googleauth/default_credentials.rb
viacheslav-rostovtsev, line 95: let's just rig it to call `determine_creds_class`. It's going to be a bit ugly but it's for the best

lib/googleauth/user_refresh.rb
```diff
@@ -39,21 +39,34 @@ class UserRefreshCredentials < Signet::OAuth2::Client
...
-        user_creds ||= {
-          "client_id"     => ENV[CredentialsLoader::CLIENT_ID_VAR],
-          "client_secret" => ENV[CredentialsLoader::CLIENT_SECRET_VAR],
-          "refresh_token" => ENV[CredentialsLoader::REFRESH_TOKEN_VAR],
-          "project_id"    => ENV[CredentialsLoader::PROJECT_ID_VAR],
-          "quota_project_id" => nil,
-          "universe_domain" => nil
-        }
+        if json_key_io
+          json_key = MultiJson.load json_key_io.read
```
viacheslav-rostovtsev, line 52: You're getting rubocop dings for method too long here.
I would recommend taking the verification you've added (in 3 classes) and see if it can be folded in `read_json` method.

viacheslav-rostovtsev (changes requested):
Let's remove the duplication in these classes while you are working in the space.

lib/googleauth/credentials.rb
aandreassa: I couldn't find any reason, so I changed it.

lib/googleauth/default_credentials.rb
aandreassa: Thanks, I refactored a bit!

lib/googleauth/external_account.rb
```diff
@@ -34,6 +34,9 @@ class Credentials
         MISSING_CREDENTIAL_SOURCE = "missing credential source for external account".freeze
         INVALID_EXTERNAL_ACCOUNT_TYPE = "credential source is not supported external account type".freeze
 
+        # The type name for this credential.
```
viacheslav-rostovtsev, line 37:
```
# @private
# @type [::String]
```

for this and all other new constants

lib/googleauth/user_refresh.rb
```diff
@@ -39,21 +39,28 @@ class UserRefreshCredentials < Signet::OAuth2::Client
...
-        user_creds = read_json_key json_key_io if json_key_io
-        user_creds ||= {
-          "client_id"     => ENV[CredentialsLoader::CLIENT_ID_VAR],
-          "client_secret" => ENV[CredentialsLoader::CLIENT_SECRET_VAR],
-          "refresh_token" => ENV[CredentialsLoader::REFRESH_TOKEN_VAR],
-          "project_id"    => ENV[CredentialsLoader::PROJECT_ID_VAR],
-          "quota_project_id" => nil,
-          "universe_domain" => nil
-        }
+        if json_key_io
```
viacheslav-rostovtsev, line 51: make it same form as in `service_account`? `user_creds = if ... else ... end`

(but please don't make the read_json_key method public)

viacheslav-rostovtsev (changes requested):

lib/googleauth/user_refresh.rb
aandreassa: Done, thanks!

lib/googleauth/credentials.rb
```diff
@@ -560,48 +568,33 @@ def duplicate options = {}
...
-      # @param [String] keyfile Path to the keyfile
-      # @raise [Google::Auth::InitializationError] If the keyfile does not exist
-      def verify_keyfile_exists! keyfile
-        exists = ::File.file? keyfile
-        raise InitializationError, "The keyfile '#{keyfile}' is not a valid file." unless exists
-      end
-
       # Initializes the Signet client.
-      def init_client hash, options = {}
+      def self.init_client hash, options = {}
```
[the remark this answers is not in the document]
aandreassa: It won't work that way even if using `protected`, so I switched it to adding `@ private`

PR #547, opened 2025-10-28
feat: Add ADC support for impersonated credentials

sai-sunder-s:
1. Add support for loading an impersonated service account credential from a JSON (typically produced by running `gcloud auth application-default login --impersoante-service-account <service account email>`)
2. Support the JSON as part of ADC flow

Sample JSON:
```
{
  "delegates": [],
  "scopes": [
    "https://www.googleapis.com/auth/drive"
  ],
  "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/service-account-email@project-name.iam.gserviceaccount.com:generateAccessToken",
  "source_credentials": {
    "client_id": "oauth_client_id",
    "client_secret": "oauth_client_secret",
    "refresh_token": "user_refresh_token",
    "type": "authorized_user",
    "universe_domain": "googleapis.com"
  },
  "type": "impersonated_service_account"
}
```

### DIFF
--- a/lib/googleauth/impersonated_service_account.rb
+++ b/lib/googleauth/impersonated_service_account.rb
@@ -92,38 +92,45 @@ module Google
       #
       # @return [Google::Auth::ImpersonatedServiceAccountCredentials]
       def self.make_creds options = {}
-        json_key_io, scope, base_credentials, source_credentials =
-          options.values_at :json_key_io, :scope, :base_credentials, :source_credentials
-
-        if json_key_io
-          if base_credentials || source_credentials
-            raise "json_key_io is not compatible with base_credentials or source_credentials"
-          end
-
-          require "googleauth/default_credentials"
-          impersonated_json = MultiJson.load json_key_io.read
-          source_credentials_info = impersonated_json["source_credentials"]
-
-          if source_credentials_info["type"] == CREDENTIAL_TYPE_NAME
-            raise "Source credentials can't be of type impersonated_service_account"
-          end
-
-          source_credentials = DefaultCredentials.make_creds(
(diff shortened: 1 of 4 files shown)


lib/googleauth/impersonated_service_account.rb
```diff
@@ -87,7 +90,32 @@ class ImpersonatedServiceAccountCredentials
       #
       # @return [Google::Auth::ImpersonatedServiceAccountCredentials]
       def self.make_creds options = {}
-        new options
+        require "googleauth/default_credentials"
```
viacheslav-rostovtsev: should be inside the if, closer to the place where we need it.

lib/googleauth/impersonated_service_account.rb
```diff
@@ -87,7 +90,32 @@ class ImpersonatedServiceAccountCredentials
...
+        json_key_io, scope = options.values_at :json_key_io, :scope
```
viacheslav-rostovtsev: this new option is not compatible to `base_credentials` or `source_credentials`. Let's have an explicit check right at the start.

lib/googleauth/impersonated_service_account.rb
viacheslav-rostovtsev: A similar `@option options` for `json_key_io` should be documented (documentation should mention that it is exclusive with `source_credentials` and `base_credentials`

viacheslav-rostovtsev (changes requested):
Please change the description. Main change is adding JSON file support to Impersonated Credentials, secondary change is enabling them in ADC flow. Scope override should not be in description .

Issue #353, opened 2021-11-15
Add full support for service account impersonation


quartzmo:
Regarding the support for service account impersonation that was discussed in #346 and that I added in #347, I noticed that googleapis/google-auth-library-java/pull/613 and golang/oauth2/pull/468 go beyond accepting a token obtained externally from [projects.serviceAccounts.generateAccessToken](https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken) and actually perform the RPC to obtain the token when provided with the appropriate service account credentials. Should this library do the same?

shpml:
Any update on this?

This is achievable with the python with the below code. 

```python
import google.auth
from google.auth import impersonated_credentials

target_scopes = ["https://www.googleapis.com/auth/cloud-platform"]
credentials, project_id = google.auth.default(scopes=target_scopes)

target_credentials = impersonated_credentials.Credentials(
          source_credentials=credentials,
          target_principal=svc_account_email,
          target_scopes=target_scopes,
          lifetime=600  # 10 mins
      )
```

konalegi:
@quartzmo do you plan to implement this feature?

konalegi:
@quartzmo I've tried to implement service account impersonation with ADC, It worked locally, although I have no idea if I'm moving right direction, since not very familiar with all this kitchen. If you have a time could you please have a look and give suggestion where to move next. 
https://github.com/googleapis/google-auth-library-ruby/pull/485/files
Thank you!

cglwn:
For others who run into this, there's the [`googleauth-extras`](https://github.com/persona-id/googleauth-extras) gem.

henrahmagix:
@viacheslav-rostovtsev Hi, thanks for implementing service account impersonation in https://github.com/googleapis/google-auth-library-ruby/pull/499!

How may I use it for Google Cloud Storage?

I've tried this but the source_credentials cannot fetch an access token 🤔 
```rb
base_credentials = Google::Cloud::Storage.default_credentials
base_credentials.fetch_access_token! # testing that default credentials works
# {"access_token"=>"<redacted>",
#  "expires_in"=>3599,
#  "token_type"=>"Bearer",
#  "id_token"=>"<redacted>",
#  "granted_scopes"=>"https://www.googleapis.com/auth/cloud-platform https://www.googleapis.com/auth/userinfo.email openid"}

[...]

PR #556, opened 2025-12-11
fix: missing auth header in impersonated service account credentials

BSick7:
This fixes an issue reporting that service account impersonation isn't working: https://github.com/googleapis/google-auth-library-ruby/issues/353.

The fixed code changes how ImpersonatedServiceAccountCredentials adds the auth header to its client. The primary issue was that the hash passed into the `apply` function is not mutated. Instead, the underlying implementations copies the hash, performs authentication, then returns the copy. The `prepare_auth_header` method was changed to align with the underlying implementation.

### DIFF
--- a/lib/googleauth/impersonated_service_account.rb
+++ b/lib/googleauth/impersonated_service_account.rb
@@ -296,9 +296,7 @@ module Google
       # @private
       # @return [Hash] The authorization header with the source credentials' token
       def prepare_auth_header
-        auth_header = {}
-        @source_credentials.updater_proc.call auth_header
-        auth_header
+        @source_credentials.updater_proc.call {}
       end
 
       # Makes the HTTP request to the impersonation endpoint.


lib/googleauth/impersonated_service_account.rb
```diff
@@ -296,9 +296,7 @@ def fetch_access_token! _options = {}
       # @private
       # @return [Hash] The authorization header with the source credentials' token
       def prepare_auth_header
-        auth_header = {}
-        @source_credentials.updater_proc.call auth_header
-        auth_header
+        @source_credentials.updater_proc.call {}
```
arenclissold, line 299: For us we needed to keep defining the `auth_header` here. The fix seems to be removing `auth_header` from the return

```rb
def prepare_auth_header
  auth_header = {}
  @source_credentials.updater_proc.call auth_header
end
```

aandreassa:
Appreciate the help here, @BSick7. We just submitted a fix as part of `googleauth` `v1.17.2` 

https://github.com/googleapis/google-auth-library-ruby/pull/587
https://rubygems.org/gems/googleauth/versions/1.17.2

Closing!

lib/googleauth/impersonated_service_account.rb
aandreassa: Please see https://github.com/googleapis/google-auth-library-ruby/pull/556#issuecomment-5062922425 for the fix.

PR #562, opened 2026-02-12
fix: return response body from revoke! for logging pipeline

### LINKED ISSUE #540: UserRefreshCredentials.revoke! block returns 0 which causes revoke to fail
There seems to be a bug in the revoke! method of Google::Auth::UserRefreshCredentials 

retry_with_error calls log_response which attempts to parse the yielded result as JSON. As the code below returns 0 it is treated as valid json and then we get a "no implicit conversion of String into Integer" error as log_response tries to use this value as a hash (response_hash["access_token"])

      def revoke!(options = {})
        c = options[:connection] || Faraday.default_connection

        retry_with_error do
          resp = c.post(REVOKE_TOKEN_URI, token: refresh_token || access_token)
          case resp.status
          when 200
            self.access_token = nil
            self.refresh_token = nil
            self.expires_at = 0
            #
            #  Need to return resp here, which fixes the crash caused by retry_with_error's use of log_response
            #
            resp
          else
            raise(Signet::AuthorizationError,
                  "Unexpected error code #{resp.status}")
          end
        end
      end

katsu223:
Fixes a bug in `UserRefreshCredentials.revoke!` where the method was returning an integer (0) instead of the response body, causing JSON parsing errors in `log_response`.

Fixes: https://github.com/googleapis/google-auth-library-ruby/issues/540

### DIFF
--- a/lib/googleauth/signet.rb
+++ b/lib/googleauth/signet.rb
@@ -210,10 +210,12 @@ module Signet
           digest = Digest::SHA256.hexdigest response_hash["id_token"]
           response_hash["id_token"] = "(sha256:#{digest})"
         end
-        Google::Logging::Message.from(
-          message: "Received auth token response: #{response_hash}",
-          "credentialsId" => object_id
-        )
+        logger&.debug do
+          Google::Logging::Message.from(
+            message: "Received auth token response: #{response_hash}",
+            "credentialsId" => object_id
+          )
+        end
       end
 
       def log_auth_error err
--- a/lib/googleauth/user_refresh.rb
+++ b/lib/googleauth/user_refresh.rb
@@ -146,6 +146,8 @@ module Google
               principal: principal
             )
           end
+
+          resp.body
         end
       end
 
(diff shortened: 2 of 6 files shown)


viacheslav-rostovtsev:
Hello! Thank you for raising the issue and implementing a fix. While looking into it I've also noticed that the `log_response` method itself never invokes the logger. I've changed that (will now log at `debug`) and pushed a commit to your branch (there are also tests for logging and execution speedups). 

LMK if you would prefer me to do this in a separate PR rather than yours

katsu223:
> Hello! Thank you for raising the issue and implementing a fix. While looking into it I've also noticed that the `log_response` method itself never invokes the logger. I've changed that (will now log at `debug`) and pushed a commit to your branch (there are also tests for logging and execution speedups).
> 
> LMK if you would prefer me to do this in a separate PR rather than yours

Hello. Thank you for taking an eye on that. I personally don't mind, you may leave changes here

aandreassa (approved):

viacheslav-rostovtsev (approved):

viacheslav-rostovtsev (approved):

PR #566, opened 2026-03-10
chore: Ruby 3.2 minimum, 4.0 default

aandreassa:
Adding `fiddle`, `irb`, `pstore` and `syslog` as they have been unbundled from Ruby 4.0.


viacheslav-rostovtsev (approved):
LGTM

PR #569, opened 2026-05-22
feat: implement Regional Access Boundary

quartzmo:
Restrict credential usage to specific regions by attaching the `x-allowed-locations` header to requests to global endpoints.

* Add internal RAB data structure and cache.
* Add lookup client (Fetcher) with retry and cooldown logic.
* Integrate into BaseClient and gate unsupported usages.
* Add URL lookup/construction for Service Account, GCE, Workload, and Workforce Identity.
* Add samples in `samples/regional_access_boundary` for demos and manual testing.


lib/googleauth/regional_access_boundary.rb
```diff
@@ -0,0 +1,35 @@
...
+    module RegionalAccessBoundary
+      autoload :RegionalAccessBoundaryData, "googleauth/regional_access_boundary/data"
+      autoload :Cache, "googleauth/regional_access_boundary/cache"
+      autoload :Fetcher, "googleauth/regional_access_boundary/fetcher"
+
+      @cache = Cache.new
+
+      # Returns the module-level cache instance.
+      #
+      # @return [Google::Auth::RegionalAccessBoundary::Cache] the cache instance.
```
aandreassa, line 29: It seems to me that `@cache` is instantiated once and shared across all credentials. Since the Cache class currently holds only a single `@data` value (without any keys), won't this cause issues in applications that use multiple credentials concurrently?

lib/googleauth/base_client.rb
```diff
@@ -85,6 +97,104 @@ def principal
...
+      #
+      # Design (Fail Open):
+      # - Run inside a separate Thread so lookup latency does not delay the primary API call.
+      # - If `regional_access_boundary_url` is nil, skip lookup (e.g. metadata server cold start).
+      # - Rescue all StandardErrors to ensure no background fetch failures propagate or crash the process.
+      #
+      # @private
+      # @param cache [Google::Auth::RegionalAccessBoundary::Cache] the cache instance.
+      # @return [Thread] the background thread instance.
+      def trigger_async_rab_fetch cache
```
aandreassa: Is it true that in `trigger_async_rab_fetch`, if the background thread gets killed or crashes on a non-`StandardError`, the cache will get stuck in `@is_fetching` = true forever?

lib/googleauth/regional_access_boundary.rb
quartzmo: Great catch! I will fix this by changing `RegionalAccessBoundary::Cache` to store entries in a thread-safe hash keyed by the `allowedLocations` lookup URL (or if GKE, the new `:unsupported` sentinel).

lib/googleauth/base_client.rb
quartzmo: Yes this seems like a very real problem. Thanks for catching it. I'll add a commit to wrap the background thread logic in an `ensure` block with a `success` tracking flag.

lib/googleauth/regional_access_boundary.rb
quartzmo: Fixed in [fcace81](https://github.com/googleapis/google-auth-library-ruby/pull/569/commits/fcace818aae0bf2c17f58d3a9e5bd65660533b89), thanks!

lib/googleauth/base_client.rb
quartzmo: Fixed in [3b26df8](https://github.com/googleapis/google-auth-library-ruby/pull/569/commits/3b26df8652646e643f22c2cfa2a899f5ecd01bfe), thanks!

spec/spec_helper.rb
```diff
@@ -42,6 +43,20 @@
 
   config.before(:each) do
     allow(Google::Auth::CredentialsLoader).to receive(:load_gcloud_project_id).and_return("my-project-id")
+    
+    # Disable Regional Access Boundary globally in tests to avoid unexpected
+    # background threads and network calls in specs that are not testing RAB.
+    [
+      Google::Auth::GCECredentials,
```
aandreassa, line 50: Should we stub `Google::Auth::BaseClient` directly? It is currently missing `PluggableAuthCredentials` AFAICT. If we can cover instead the base module it would make future-proof.

lib/googleauth/base_client.rb
```diff
@@ -85,6 +97,117 @@ def principal
...
+      end
+
+      # Determines if a request is eligible for Regional Access Boundary restrictions.
+      #
+      # @private
+      # @param opts [Hash] request options.
+      # @return [Boolean] true if the header should be applied, false otherwise (failing open).
+      def should_apply_rab? opts
+        # Return early if not supported by credential type or if ID token.
+        return false unless token_type != :id_token && supports_regional_access_boundary?
```
aandreassa: It is a bit hard to read. Based on the comment above, I think we should write as

`return false if token_type == :id_token || !supports_regional_access_boundary?` 

instead.

lib/googleauth/regional_access_boundary/cache.rb
```diff
@@ -0,0 +1,193 @@
...
+# limitations under the License.
+
+require "monitor"
+
+module Google
+  module Auth
+    module RegionalAccessBoundary
+      # Cache stores and manages the lifecycle of Regional Access Boundary data.
+      #
+      # Concurrency & Multi-Tenancy Design:
```
aandreassa, line 22: Great docs!

samples/regional_access_boundary/default_universe.rb
```diff
@@ -0,0 +1,92 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "googleauth"
+require "faraday"
+require "multi_json"
```
aandreassa: Can we do without the `multi_json`?

spec/googleauth/regional_access_boundary/integration_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+  let(:key) { client.regional_access_boundary_url }
+
+  before do
+    # Stub the module-level cache to use our isolated test cache
+    allow(Google::Auth::RegionalAccessBoundary).to receive(:cache).and_return(cache)
+    # Stub Thread.new to yield immediately, running the fetch in the main thread
+    # to avoid WebMock thread-safety issues and ensure predictable test execution.
+    allow(Thread).to receive(:new).and_yield
+
+    # Re-enable RAB lookup for the test client (overriding global disable in spec_helper)
```
aandreassa, line 82: 👍

aandreassa (commented):
LG @quartzmo! I will take another good look tomorrow, but a few nits for now

samples/regional_access_boundary/default_universe.rb
quartzmo: Great catch! Of course. This PR is showing its age.

quartzmo: Merged main into branch and updated to json, thanks!

lib/googleauth/regional_access_boundary/cache.rb
quartzmo: Thanks!! Concurrency management should always extra comments if you ask me.

spec/spec_helper.rb
quartzmo: Good idea! Done, thanks!

lib/googleauth/base_client.rb
quartzmo: Done, thanks!

aandreassa (approved):

quartzmo:
Adding `do not merge` pending approval by @nbayati.

nbayati (approved):
Thank you @quartzmo for the PR. The implementation looks good to me, the only comment I have is regarding the pattern matching for service account emails. I updated the recommendation in go/sdk-rab-mwlid to check for a standard email pattern (e.g., ^[^@]+@[^@]+\.[^@]+$). 
However, we might need to make it even stricter, at least python (See https://github.com/googleapis/google-cloud-python/issues/17711). I don't know yet if this is a gcloud specific issue, or if we should make it more strict across the board.

PR #580, opened 2026-06-12
test(auth): enforce strict once call counts on token caching tests

westarle:
The Ruby specs relied on loose have_been_requested expectations which only check if a request happened, not how many times. Other languages (like Python, Swift, and Java) verify caching logic by asserting that exactly one outbound HTTP request is made regardless of the number of credential calls.


quartzmo (approved):
Thank you for catching these!

quartzmo:
Trying to get the remaining GH Action presubmits to run.

PR #581, opened 2026-06-15
test(auth): verify unsupported types from ADC

westarle:
Add unit tests to verify that when Application Default Credentials (ADC) is initialized, if it loads an unsupported credential JSON payload (such as id_token, service_account_signer, or authorized_user missing refresh token) through the environment or well-known paths, it successfully rejects it and raises a Google::Auth::InitializationError.

Unlike other client libraries (which might use different exception types), the Ruby library consistently raises a Google::Auth::InitializationError.

### DIFF
--- a/spec/googleauth/get_application_default_spec.rb
+++ b/spec/googleauth/get_application_default_spec.rb
@@ -22,7 +22,11 @@ require "googleauth"
 require "spec_helper"
 require "os"
 
+include Google::Auth
+include Google::Auth::CredentialsLoader
+
 describe "#get_application_default" do
+
   # Pass unique options each time to bypass memoization
   let(:options) { |example| { dememoize: example } }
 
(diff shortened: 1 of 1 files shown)


spec/googleauth/get_application_default_spec.rb
```diff
@@ -324,4 +328,45 @@ def cred_json_text
...
+    let(:unsupported_types) do
+      [
+        { type: "id_token" },
+        { type: "service_account_signer" },
+        { type: "authorized_user" } # missing client_id, client_secret, refresh_token
+      ]
+    end
+
+    it "fails if the GOOGLE_APPLICATION_CREDENTIALS file contains the unsupported creds" do
+      unsupported_types.each do |cred_json|
```
quartzmo, line 342: If one type fails, RSpec will abort the loop and won't check the remaining types. Why not just explicitly test each type in its own `it` block, and maybe refactor to a helper to DRY it up?

spec/googleauth/get_application_default_spec.rb
```diff
@@ -324,4 +328,45 @@ def cred_json_text
...
+          ENV[@var_name] = key_path
+          expect do
+            Google::Auth.get_application_default @scope, options
+          end.to raise_error Google::Auth::InitializationError
+        end
+      end
+    end
+
+    it "fails if the well known file contains the unsupported creds" do
+      unsupported_types.each do |cred_json|
```
quartzmo, line 356: Same as above, test each type in its own it block.

quartzmo (changes requested):

PR #586, opened 2026-07-23
fix: return auth header from source_credentials in ImpersonatedServiceAccountCredentials

### LINKED ISSUE #563: ADC support for impersonated credentials failing with 401 - Reason CREDENTIALS_MISSING
The actual request, after generating an auth token, is missing the auth headers. The code tries to [prepare the auth header using the source credentials, **presuming the passed in hash will be mutated**](https://github.com/googleapis/google-auth-library-ruby/blob/5c4ab1cfc3517d5490633c878cb5136ca4da75e6/lib/googleauth/impersonated_service_account.rb#L298C7-L302C10).

However, `source_credentials` subclassing from BaseClient will [**return a modified copy of the passed in hash** rather than mutating the passed in hash](https://github.com/googleapis/google-auth-library-ruby/blob/5c4ab1cfc3517d5490633c878cb5136ca4da75e6/lib/googleauth/base_client.rb#L45C8-L49C10
).

aandreassa:
In `ImpersonatedServiceAccountCredentials#prepare_auth_header`, `@source_credentials.updater_proc.call` returns a new hash with authorization headers rather than mutating in-place. Discarding its return value caused outgoing IAM impersonation requests to miss the `Authorization` header and fail with `401 CREDENTIALS_MISSING`.

This change updates `prepare_auth_header` to return the hash from `updater_proc.call({})` and updates unit specs to verify header presence.

Fixes #563

### DIFF
--- a/lib/googleauth/impersonated_service_account.rb
+++ b/lib/googleauth/impersonated_service_account.rb
@@ -297,9 +297,9 @@ module Google
       # @private
       # @return [Hash] The authorization header with the source credentials' token
       def prepare_auth_header
-        auth_header = {}
-        @source_credentials.updater_proc.call auth_header
-        auth_header
+        # updater_proc returns a new hash containing authorization headers
+        # rather than mutating the passed hash in place.
+        @source_credentials.updater_proc.call({})
       end
 
       # Makes the HTTP request to the impersonation endpoint.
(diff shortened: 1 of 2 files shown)


quartzmo (approved):
This fix LGTM. Thank you!

PR #588, opened 2026-07-23
fix(auth): validate options parameter in ServiceAccountCredentials.make_creds

### LINKED ISSUE #482: googleauth-1.11.0/lib/googleauth/service_account.rb:87:in `unescape': undefined method `gsub' for nil:NilClass
- OS: ubuntu 22.04
  - Ruby version: 2.7.0
  - Gem name and version: googleauth 1.11.0

#### Steps to reproduce

  1. Attempt to use make_creds with json_token

#! /usr/bin/ruby

require 'googleauth'

SERVICE_ACCOUNT_PRIVATE_KEY = "-----BEGIN PRIVATE KEY-----MIIE1234ABCD-----END PRIVATE KEY-----" 

begin
  credentials = Google::Auth::ServiceAccountCredentials.make_creds(
    scope: "https://analyticsreporting.googleapis.com/v4/reports:batchGet", 
    json_key: SERVICE_ACCOUNT_PRIVATE_KEY
  )
rescue => error
  puts "Error: #{error.message}"
end

### LINKED ISSUE #544: auth
A service account is created and the spreadsheet is shared with the service account e-mail. However when using sheetsV4 as below the error that comes back is

`forbidden: Method doesn't allow unregistered callers (callers without established identity)`

This makes no sense because the service account is basically its own identity. It is completely unclear how this "identity" is determined or how we can establish an identity to this service account, whatever that means. This may be an issue with the way the library code is calling the API because this sometimes works on other google accounts.

The code:

```ruby
require 'google/apis/sheets_v4'
require 'googleauth'

SCOPES = [
	'https://www.googleapis.com/auth/drive',
	'https://www.googleapis.com/auth/spreadsheets'
]

auth = Google::Auth::ServiceAccountCredentials.make_creds(
		json_key_io: File.open('service_account.json'),
		scopes: SCOPES
	)

sheet = Google::Apis::SheetsV4::SheetsService.new
sheet.authorization = auth

p sheet.get_spreadsheet('1T7kwW5-Hu1idYh1I6CpqgRHUj3SWCtVLP2qljTFL3hs')
```

aandreassa:
- Warn on unrecognized options passed to `ServiceAccountCredentials.make_creds`.
- Raise `InitializationError` when required credential fields or valid `json_key_io` objects are missing.
- Fix `unescape(nil)` bug to handle `nil` inputs safely.
- Update `make_creds` docstring to document all supported options.

Fixes #482
Fixes #544

### DIFF
--- a/lib/googleauth/service_account.rb
+++ b/lib/googleauth/service_account.rb
@@ -45,6 +45,19 @@ module Google
       # @type [::String] The type name for this credential.
       CREDENTIAL_TYPE_NAME = "service_account".freeze
 
+      # @private
+      # @type [Array<Symbol>] Allowed option keys for make_creds.
+      VALID_MAKE_CREDS_OPTIONS = [
+        :json_key_io,
+        :scope,
+        :enable_self_signed_jwt,
+        :target_audience,
+        :audience,
+        :token_credential_uri,
+        :default_connection,
+        :connection_builder
+      ].freeze
+
       def enable_self_signed_jwt?
         # Use a self-singed JWT if there's no information that can be used to
         # obtain an OAuth token, OR if there are scopes but also an assertion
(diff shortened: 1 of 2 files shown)


spec/googleauth/service_account_spec.rb
```diff
@@ -125,9 +125,51 @@ def cred_json_text_with_universe_domain
...
-    )
+    end.not_to raise_error
+  end
+
+  it "raises InitializationError when credentials are missing" do
+    expect do
+      ServiceAccountCredentials.make_creds(
+        scope: "https://www.googleapis.com/auth/userinfo.profile"
+      )
+    end.to raise_error(Google::Auth::InitializationError, /Missing required field/)
```
torreypayne, line 136: 🥇

lib/googleauth/service_account.rb
```diff
@@ -212,7 +243,20 @@ def self.creds_from_env
         [private_key, client_email, project_id, nil, nil]
       end
 
-      private_class_method :creds_from_env
+      # @private
+      # Validates options passed to make_creds and emits a warning for unrecognized keys.
+      #
+      # @param options [Hash] The options hash to validate.
+      def self.validate_make_creds_options options
+        return unless options.is_a? Hash
```
torreypayne, line 251: Nit: based on the feedback from issue #482 would it make sense to add a warn or raise an error on some of the input types? There seemed to be confusion around String vs IO input for `json_key` and `json_key_io` so I'm wondering if we can clarify the typing a bit more.

torreypayne (approved):

lib/googleauth/service_account.rb
aandreassa: There is a `warn` just down below on using incorrect entry names (most common mistake). I also defined the types in the docstrings. 

`json_key_io.respond_to? :read` above is to check `IO` vs `String`, which will raise an error.

Do you think that's sufficient?

PR #589, opened 2026-07-24
docs: Update CLI/installed app example to use loopback IP redirect

### LINKED ISSUE #446: Installed app sample
Can someone tell me why [Example (Command Line) [Deprecated]](https://github.com/googleapis/google-auth-library-ruby#example-command-line-deprecated) is deprecated.

Yes oob has stopped working but all of the other client libraries,  .net, php, python, golang, java have just switched to using http:\\localhost as the migration guide recommends [oob-migration](https://developers.google.com/identity/protocols/oauth2/resources/oob-migration)

Any special reason why that wont work in this library and with the existing sample?

aandreassa:
Update the CLI example in README.md to use the loopback IP redirect flow (`http://localhost`) instead of the deprecated OOB URI (`urn:ietf:wg:oauth:2.0:oob`), as recommended in the Google OOB Migration Guide.

Fixes #446

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,30 +145,30 @@ get('/oauth2callback') do
 end
 ```
 
-### Example (Command Line) [Deprecated]
+### Example (Command Line / Installed App)
 
-The Google Auth OOB flow has been discontiued on January 31, 2023. The OOB flow is a legacy flow that is no longer considered secure. To continue using Google Auth, please migrate your applications to a more secure flow. For more information on how to do this, please refer to this [OOB Migration](https://developers.google.com/identity/protocols/oauth2/resources/oob-migration) guide.
+Command line and desktop applications should use the Loopback IP redirect flow (`http://localhost`) as recommended in the [OOB Migration Guide](https://developers.google.com/identity/protocols/oauth2/resources/oob-migration).
 
 ```ruby
-require 'googleauth'
-require 'googleauth/stores/file_token_store'
+require "googleauth"
+require "googleauth/stores/file_token_store"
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -145,30 +145,30 @@ get('/oauth2callback') do
...
+authorizer = Google::Auth::UserAuthorizer.new client_id, scope, token_store
 
-user_id = ENV['USER']
-credentials = authorizer.get_credentials(user_id)
+user_id = ENV["USER"]
+credentials = authorizer.get_credentials user_id
 if credentials.nil?
-  url = authorizer.get_authorization_url(base_url: OOB_URI )
+  url = authorizer.get_authorization_url base_url: LOOPBACK_URI
   puts "Open #{url} in your browser and enter the resulting code:"
```
quartzmo:
```suggestion
    puts "Open this URL in your browser:\n#{url}\n\n" \
         "After authorizing, your browser will redirect to a localhost URL (which " \
         "will likely display a connection error). Copy the 'code' parameter value " \
         "from the browser's address bar and enter it here:"
```

quartzmo (changes requested):
The `puts` text seems a bit ambiguous, can we help the user better understand what to expect? Here's a suggestion, feel free to tweak it.

README.md
aandreassa: Thank you for the suggestion! I decided to split it into a few `puts` statements to make it easier to follow as well. PTAL

torreypayne (approved):

quartzmo (approved):

